Issue #1, opened 2016-02-29
Create and update yml files


roigcarlo:
The following yml files need to be updated according to: https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/SimPhoNy_Metadata_Schema
- Cuba.yml
- cuds_item.yml
- material_relation_definitions.yml

Simphony.yml needs to be split into 
- Physics Equations
- Material Relations
- Materials
- Boundaries
- Conditions

tuopuu:
It seems that simphony.yml is not going to be that large and complex file that it really needs to be split into 5 separate files. 

I think that the main idea supporting splitting we had in the dev-team meeting was that SimPhoNy classes described in this file would be easier to update, and it would be easier to define new classes and place them in correct "categories" in these files. However, updating or adding new stuff is not going to happen so often. Once we roll out version 1.0, that should cover every Model/Engine we have in SimPhoNy at the moment. Also, version control will be straightforward when everything is in a single file. I think we should refrain from splitting simphony.yml for now, until there is a real need for it. @ahashibon, do you have some recommendation to this issue?

ahashibon:
Agree!

mehdisadeghi:
Having less files is definitely better for long term. In the beginning we can focus on two files, one for CUBA one for CUDS.

kitchoi:
I agree with having less files.

tuopuu:
cuba.yml was updated to the new yaml format. material_relations.yml was merged to simphony_metadata.yml in #7

roigcarlo:
Implemented in #7

PR #3, opened 2016-03-08
new scheme according to the final SSB proposal from 3.4.2016

ahashibon:
@khiltunen @kemattil @pooyan-dadvand @simphony/cimne @simphony/fraunhofer-iwm @simphony/enthought @SGGgarcia and Guillermo

Please review the new schema according to the SSB proposal. 
It is an initial version, more relations, elements, etc are needed.

### DIFF
--- a/simphony_metadata.yml
+++ b/simphony_metadata.yml
@@ -94,7 +94,7 @@ CUDS_KEYS:
     parent: CUBA.COMPUTATIONAL_MODEL
  
   MODEL_EQUATION:
-    definition: SimPhoNy Model Equation = PE + MRs according to the RoMM V4. 
+    definition: SimPhoNy Model Equation = PE + MRs according to the RoMM
     models: []
     parent: CUBA.CUDS_COMPONENT
     physics_variables: []
@@ -170,7 +170,7 @@ CUDS_KEYS:
     definition: Classical Atomistic Molecular Dynamics using Newtons equations of motion
     parent: CUBA.PHYSICS_EQUATION
     models: [CUBA.ATOMISTIC]
-    physics_variables: [CUBA.POSITION, CUBA.VELOCITY, CUBA.MOMENTUM, CUBA.ACCELERATION, CUBA.FORCEÃ¤]  
+    physics_variables: [CUBA.POSITION, CUBA.VELOCITY, CUBA.MOMENTUM, CUBA.ACCELERATION, CUBA.FORCE]  
 
   MOLECULAR_STATICS:
     definition: Classical Atomistic static molecular model
@@ -225,7 +225,7 @@ CUDS_KEYS:
   # note:GRAVITY and CUBA.ELECROSTATIC_FIELD are defined in cuba.yml. 
 
   ISOTHERMAL_MODEL:
-    definition: Isothermal heat transport model, no transport of heat is assumed. 
+    definition: Isothermal heat transport model, no transport of heat is assumed
     parent: CUBA.PHYSICS_EQUATION
     models: [CUBA.CONTINUUM]
     physics_quantities: []
@@ -239,7 +239,7 @@ CUDS_KEYS:
   INCOMPRESSIBLE_FLUID_MODEL:
     parent: CUBA.PHYSICS_EQUATION
     models: [CUBA.CONTINUUM]
-    definition: Incompressible fluid model.
+    definition: Incompressible fluid model
     physics_quantities: [CUBA.VELOCITY, CUBA.POSITION, CUBA.DENSITY, CUBA.VICOSITY]
     
   LAMINAR_FLOW_MODEL:
@@ -253,13 +253,13 @@ CUDS_KEYS:
     models: [CUBA.CONTINUUM]
 
   FREE_SURFACE_MODEL:
-    definition: Free surface model. 
+    definition: Free surface model
     parent: CUBA.PHYSICS_EQUATION
     models: [CUBA.CONTINUUM]
     physics_quantities: [CUBA.SURFACE_TENSION]
     
   MIXTURE_MODEL:
-    definition: mixture (drift flux) model.
+    definition: mixture (drift flux) model
     parent: CUBA.PHYSICS_EQUATION
     models: [CUBA.CONTINUUM]
 
@@ -270,12 +270,12 @@ CUDS_KEYS:
     default: CUBA.NEWTONIAN_FLUID_MODEL
 
   NEWTONIAN_FLUID_MODEL:
-    definition: Newtonian ﬂuid model assuming the viscous stresses are proportional to the rates of deformation. 
+    definition: Newtonian ﬂuid model assuming the viscous stresses are proportional to the rates of deformation
     parent: CUBA.RHEOLOGY_MODEL
     models: [CUBA.CONTINUUM]
   
   POWER_LAW_VISCOSITY_MODEL:
-    definition: Power Law model for a variable viscosity function that is limited by minimum and maximum values. 
+    definition: Power Law model for a variable viscosity function that is limited by minimum and maximum values 
     parent: CUBA.RHEOLOGY_MODEL
     models: [CUBA.CONTINUUM]
     CUBA.MAXIMUM_VISCOSITY: 
@@ -288,7 +288,7 @@ CUDS_KEYS:
       default: 1
       
   HERSCHEL_BULKLEY_MODEL:
-    definition: Herschel-Bulkley model combines the effects of Bingham plastic and power-law behavior in a fluid.
+    definition: Herschel-Bulkley model combines the effects of Bingham plastic and power-law behavior in a fluid
     parent: CUBA.RHEOLOGY_MODEL
     models: [CUBA.CONTINUUM]
     CUBA.INITIAL_VISCOSITY: 
@@ -338,7 +338,7 @@ CUDS_KEYS:
       shape: [2]
 
   DIRICHLET_BOUNDARY_CONDITION:
-    definition: Dirichlet boundary condition.
+    definition: Dirichlet boundary condition
     parent: CUBA.CONDITION    
     models: [CUBA.CONTINUUM]
     CUBA.PHYSICS_VARIABLE: 
@@ -347,7 +347,7 @@ CUDS_KEYS:
       shape: [1,]
 
   NEUMANN_BOUNDARY_CONDITION:
-    definition: Neumann boundary condition.
+    definition: Neumann boundary condition
     parent: CUBA.CONDITION
     models: [CUBA.CONTINUUM]
     CUBA.PHYSICS_VARIABLE: 
@@ -357,7 +357,7 @@ CUDS_KEYS:
 
   EMPTY_BOUNDARY_CONDITION:
     parent: CUBA.CONDITION
-    definition: Empty boundary condition.
+    definition: Empty boundary condition
     models: [CUBA.CONTINUUM]
     CUBA.PHYSICS_VARIABLE: 
       shape: [1,]
@@ -376,7 +376,7 @@ CUDS_KEYS:
   
   LENNARD_JONNES_6_12:
     parent: CUBA.PAIR_POTENTIAL
-    definition: A Lennard Jonnes 6-12 Potential. 
+    definition: A Lennard Jonnes 6-12 Potential 
     models: [CUBA.ATOMISTIC]
     physics_variables: [CUBA.POSITION, CUBA.POTENTIAL_ENERGY]
     CUBA.CUTOFF_DISTANCE:
@@ -388,7 +388,7 @@ CUDS_KEYS:
 
   COULOMB:
     parent: CUBA.PAIR_POTENTIAL
-    definition: The standard electrostatic Coulombic interaction potential between a pair of point charges.
+    definition: The standard electrostatic Coulombic interaction potential between a pair of point charges
     models: [CUBA.ATOMISTIC]
     CUBA.DIELECTRIC_CONSTANT:
       default: 1.0
--- a/simphony_metadata.yml
+++ b/simphony_metadata.yml
@@ -139,7 +139,7 @@ CUDS_KEYS:
     models: [CUBA.MESOSCOPIC, CUBA.CONTINUUM]
     physics_variables: [CUBA.ACCELERATION]
     CUBA.ACCELERATION:
-      default: [0, 0, 0]
+      default: [0., 0., 0.]
 
   CONSTANT_ELECTROSTATIC_FIELD_MODEL:
     definition: A constant elecrostatic field model 
@@ -147,7 +147,7 @@ CUDS_KEYS:
     models: [CUBA.MESOSCOPIC, CUBA.CONTINUUM]
     physics_variables: [CUBA.ELECTRIC_FIELD, CUBA.CHARGE]
     CUBA.ELECTROSTATIC_FIELD:
-      default: [0, 0, 0]
+      default: [0., 0., 0.]
   # note:GRAVITY and CUBA.ELECROSTATIC_FIELD are defined in cuba.yml. 
 
   ISOTHERMAL_MODEL:
@@ -207,7 +207,7 @@ CUDS_KEYS:
     CUBA.MAXIMUM_VISCOSITY: 
       default: 1e-3
     CUBA.MINIMUM_VISCOSITY:
-      default: 1E-5
+      default: 1e-5
     CUBA.LINEAR_CONSTANT:
       default: 1e-5
     CUBA.POWER_LAW_INDEX:
@@ -220,7 +220,7 @@ CUDS_KEYS:
     CUBA.INITIAL_VISCOSITY: 
       default: 1e-3
     CUBA.RELAXATION_TIME: 
-      default: 1
+      default: 1.0
     CUBA.LINEAR_CONSTANT:
       default: 1e-5
     CUBA.POWER_LAW_INDEX:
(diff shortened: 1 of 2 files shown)


tuopuu:
@ahashibon, there seems to be some overlap between cuba.yml and material_relation_definitions.yml with simphony_metadata.yml. Some definitions are currently present in two of these files. Are we supposed to add to simphony_metadata.yml the contents of material_relation_definitions.yml, so that we can remove that specific file completely. Is this what we're aiming to do?  

Why should we have a separate file cuds_item.yml for low level Cuds data objects? Shouldn't that be also merged with simphony_metadata.yml? Also, what are the final YAML-files we end up using? Just simphony-metadata.yml and cuba.yml?

Because there seems to be so many definitions that need to be added to simphony_metadata.yml, should we (the developers) work in parallel updating this branch before finally merging it to simphony-metadata master? Another option would be to merge this quickly, and then start working individually with separate branches that add new definitions.

mehdisadeghi:
IMO one file for CUDS and one for CUBA are enough. We might end up splitting specification into multiple files in order to maintain them effectively however.

tuopuu:
@mehdisadeghi, please take a look at mine and Adham's comment at #1. I think we can start with just two files for now, and if necessary, later split them into smaller files.

mehdisadeghi:
@tuopuu I agree.

ahashibon:
The goal would be to have one source for all metadata (simphony_metadata.yml), however right now the cuba.yml has very elemental metadata. It will take time to merge them, so my suggestion is also to leave it as is, and comment-out or better ignore those in the simphony_metadata.yml file that are also in cuba.yml for the mean time. The same for CUDS_item items. 

The materials relations should be however moved to the new metadata file, this is the decision of the SSB already at the M24 meeting. We already moved most of them.

ahashibon:
@tuopuu what final YAML-files we end up using? Just simphony-metadata.yml and cuba.yml?

This is another plausible solution, if we define the elementary cuba (in cuba.yml) using a differnet schema (the cuba are the elements that are simple, and are used across the board) while the simphony_metadata are the more complex data objects that are domain specific. This is a matter of ongoing research also in the wider level of the cluster.

tuopuu:
Thanks @ahashibon for clarification. How do we continue to work on this, as it's related to #1? Do you mind if I take over this branch?

ahashibon:
@kemattil @khiltunen @gromansg @pooyan-dadvand @tuopuu @mehdisadeghi can this be merged ?

simphony_metadata.yml
```diff
@@ -0,0 +1,412 @@
...
+    physics_variables: [CUBA.POSITION, CUBA.POTENTIAL_ENERGY]
+    CUBA.CUTOFF_DISTANCE:
+      default: 1.0
+    CUBA.ENERGY_WELL_DEPTH:
+      default: 1.0
+    CUBA.VAN_DER_WAALS_RADIUS:
+      default: 1.0
+
+  COULOMB:
+  parent: CUBA.PAIR_POTENTIAL
```
tuopuu: This should be indented.

mehdisadeghi:
> my suggestion is also to leave it as is, and comment-out or better ignore those in the simphony_metadata.yml

I would suggest to merge them into the new files (CUDS & CUBA) and delete the old ones afterwards, otherwise it will be confusing.

tuopuu:
@ahashibon, just one minor comment on the file. I also think that we should pick a convention for the number of empty lines between definitions of the objects: currently there seems to be a variation between 1-3. I propose using always one empty line.

tuopuu:
Some of the definitions seem to have a period at the end. I think we should also agree on a convention that period is only used for definitions that are clearly sentences, and leave the other "header-like" definitions without it.

simphony_metadata.yml
```diff
@@ -0,0 +1,412 @@
...
+      default: 1e-3
+    CUBA.RELAXATION_TIME: 
+      default:1
+    CUBA.LINEAR_CONSTANT:
+      default: 1e-5
+    CUBA.POWER_LAW_INDEX:
+      default: 1
+
+  CROSS_POWER_LAW_MODEL:
+    definitoin: Viscosity Cross power law model
```
tuopuu: this should be definition.

kemattil:
Please do not merge yet. I have some comments, put not yet time to list them.
Will do it tomorrow or Friday at the latest ...

ahashibon:
@tuopuu Thanks, good comments, please check now if the dots and spaces are ok.

ahashibon:
few more minor formatting.

ahashibon:
closed by mistake! sorry about that. Working through a very slow network.

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+
+  GRANULAR_DYNAMICS:
+    definition: Granular dynamics of spherical particles using DEM
+    parent: CUBA.PHYSICS_EQUATION    
+    models: [CUBA.MESCOCOPIC]
+    physics_variables: [CUBA.POSITION, CUBA.VELOCITY, CUBA.MOMENTUM, CUBA.ACCELERATION, CUBA.MOMENT_INERTIA, CUBA.TORQUE, CUBA.ANGULAR_VELOCITY]  
+    
+  CFD:
+    definition: Computational fluid dynamics general (set of ) equations for momentum, mass and energy
+    parent": CUBA.PHYSICS_EQUATION
```
tuopuu: Remove "

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+      scope: CUBA.USER
+    CUBA.TURBULENCE_MODEL:
+      default: CUBA.LAMINAR_FLOW_MODEL
+    CUBA.MULTIPHASE_MODEL: 
+      default: CUBA.SINGLE_PHASE_MODEL
+    CUBA.RHEOLOGY_MODEL:
+      default: CUBA.NEWTONIAN_MODEL
+    CUBA.GRAVITY_MODEL:
+    CUBA.ELECTROSTATIC_MODEL:
+      defualt: CUBA.CONSTANT_ELECTROSTATIC_FIELD_MODEL
```
tuopuu: Change to default

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+    CUBA.ACCELERATION:
+      default: [0, 0, 0]
+
+  CONSTANT_ELECTROSTATIC_FIELD_MODEL:
+    definition: A constant elecrostatic field model 
+    parent: CUBA.PHYSICS_EQUATION
+    models: [CUBA.MESOSCOPIC, CUBA.CONTINUUM]
+    physics_variables: [CUBA.ELECTRIC_FIELD, CUBA.CHARGE]
+    CUBA.ELECTROSTATIC_FIELD:
+      default: [0, 0, 0.0]
```
tuopuu: Could use just [0, 0, 0] here.

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+    definition: Non Isothermal heat transport model
+    parent: CUBA.PHYSICS_EQUATION
+    models: [CUBA.CONTINUUM]
+    physics_quantities: [CUBA.TEMPERATURE, CUBA.HEAT_CONDUCTIVITY]
+      
+  INCOMPRESSIBLE_FLUID_MODEL:
+    parent: CUBA.PHYSICS_EQUATION
+    models: [CUBA.CONTINUUM]
+    definition: Incompressible fluid model
+    physics_quantities: [CUBA.VELOCITY, CUBA.POSITION, CUBA.DENSITY, CUBA.VICOSITY]
```
tuopuu: Change the last key to CUBA.VISCOSITY

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+    
+  LAMINAR_FLOW_MODEL:
+    parent: CUBA.PHYSICS_EQUATION
+    definition: Laminar model.
```
tuopuu: Remove period

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+    models: [CUBA.CONTINUUM]
+  
+  POWER_LAW_VISCOSITY_MODEL:
+    definition: Power Law model for a variable viscosity function that is limited by minimum and maximum values 
+    parent: CUBA.RHEOLOGY_MODEL
+    models: [CUBA.CONTINUUM]
+    CUBA.MAXIMUM_VISCOSITY: 
+      default: 1e-3
+    CUBA.MINIMUM_VISCOSITY:
+      DEFAULT:1E-5
```
tuopuu: change to lower case: default: 1e-5

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+      default: 1
+      
+  HERSCHEL_BULKLEY_MODEL:
+    definition: Herschel-Bulkley model combines the effects of Bingham plastic and power-law behavior in a fluid
+    parent: CUBA.RHEOLOGY_MODEL
+    models: [CUBA.CONTINUUM]
+    CUBA.INITIAL_VISCOSITY: 
+      default: 1e-3
+    CUBA.RELAXATION_TIME: 
+      default:1
```
tuopuu: add space

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+      default: 1
+
+  CROSS_POWER_LAW_MODEL:
+    definition: Viscosity Cross power law model
+    parent: CUBA.RHEOLOGY_MODEL
+    models: [CUBA.CONTINUUM]
+    CUBA.INITIAL_VISCOSITY: 
+      default: 1e-3
+    CUBA.MAXIMUM_VISCOSITY: 
+      DEFAULT:1E-5   # nu_infiniy
```
tuopuu: Also, lower case here: default: 1e-5

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+      default: 0.5
+
+  BIRD_CARREAU_MODEL:
+    definition: The Bird-Carreau model
+    parent: CUBA.RHEOLOGY_MODEL
+    models: [CUBA.CONTINUUM]
+    CUBA.INITIAL_VISCOSITY: 
+      default: 1e-3
+    CUBA.MAXIMUM_VISCOSITY: 
+      DEFAULT:1E-5
```
tuopuu: Lower-case.

tuopuu:
@ahashibon, some minor typos and formatting errors, but apart from that, looks good to me.

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+# If omitted, no default is assumed, and the user has to supply a
+# value.
+
+# scope: CUBA.USER or CUBA.SYSTEM.  if omitted, the CUBA.USER is
+# assumed by default
+
+# shape: as [i,j,] specifying that it is a sequence of entities of its
+# own type.  [i, j] means an array of i*j types.  [i, ] means the
+# sequence or array can have minimum of i elements. [, i] the array can
+# have any, maximum j elements. If omitted, a shape of [1] is assumed.
```
kitchoi: > [i, ] means the sequence or array can have minimum of i elements, [, i] the array can have any, maximum j elements.

First of all, typo for the second part of the sentence (i and j).  Second of all, having this minimum and maximum flexibility is very strange.  Where is it needed?  And there is `shape: [1:]` further down this file.  What does that mean?

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+  PHYSICS_EQUATION:
+   definition: SimPhoNy Physics Equation
+   parent: CUBA.MODEL_EQUATION
+
+  MATERIAL_RELATION_EQUATION:
+    definition: SimPhoNy Material Relation
+    parent: CUBA.MODEL_EQUATION
+    CUBA.MATERIAL:
+      scope: CUBA.USER
+      shape: [1,]
```
kitchoi: Why is it necessary to have `shape: [1,]` here while other places have simply `[]`?

simphony_metadata.yml
```diff
@@ -0,0 +1,398 @@
...
+      shape: [2]
+
+  DIRICHLET_BOUNDARY_CONDITION:
+    definition: Dirichlet boundary condition
+    parent: CUBA.CONDITION    
+    models: [CUBA.CONTINUUM]
+    CUBA.PHYSICS_VARIABLES: 
+      shape: [1:]
+    CUBA.MATERIAL: 
+      shape: [1:]
```
kitchoi: What does this mean?

kitchoi:
Except for a few questions I have above, the rest looks good to me.

simphony_metadata.yml
ahashibon: Thanks for noting the typo. The notation for the list may not be the best; this should be seen as an intermediate solution. There is a need to define values for some entities that are lists/sequence/series of a CUBA.  I first used a notation with a comma, but I found it later to be a bit inconsistent with standard notation, so I changed it to use colons and mixed up things a bit apparently.  Here is how the comment should look like (will be pushed shortly: 

`# shape: [i_min:i_max] specifies that the value is a sequence of entities of its own type.`
`# In each dimension, the first index is the minimum while the second maximum number of items allowed.`
`# similarly:`
`# shape: [i_min:i_max,j_min:j_max] specifies that the value is a multidimensional`
`# sequence of entities of its own type.  In each dimension, the first index`
`# is the minimum while the second maximum number of items allowed.`

Basically, we need a way to convey the information that the value is a list/array/sequence. This may still not be the best solution. Any ideas to enhance this are welcome.

simphony_metadata.yml
ahashibon: This is meant to convey that a material relation needs to have at least one material. Note that is should be `[1:]`.

simphony_metadata.yml
ahashibon: Please see previous comment.

kitchoi:
The new push looks good to me!  Thanks @ahashibon!

tuopuu:
Ok to merge?

ahashibon:
will first create the initial Readme file with all the comments, and then it will be ok to merge.

tuopuu:
@ahashibon, great! If you'll do a new push (for the ReadMe file) I can give it a quick review.

ahashibon:
@tuopuu  please check it! Thanks!

simphony_metadata.yml
```diff
@@ -0,0 +1,415 @@
...
+      default: 1
+      
+  HERSCHEL_BULKLEY_MODEL:
+    definition: Herschel-Bulkley model combines the effects of Bingham plastic and power-law behavior in a fluid
+    parent: CUBA.RHEOLOGY_MODEL
+    models: [CUBA.CONTINUUM]
+    CUBA.INITIAL_VISCOSITY: 
+      default: 1e-3
+    CUBA.RELAXATION_TIME: 
+      default: 1
```
kitchoi: Where is the unit of relaxation time defined?

kitchoi: I am mainly interested in knowing whether it must be an integer.

ahashibon: It is not an integer ! 

SimPhoNy uses, currently, by default the SI system (MKS...). 
units should be introduced later. 
many new CUBA will have to be defined now in cuba.yml...

kitchoi: Thanks.  Then it would be better to use `1.` instead of `1`, especially important for systems still using Python2.x

simphony_metadata.yml
```diff
@@ -0,0 +1,326 @@
...
+    models: [CUBA.CONTINUUM]
+  
+  POWER_LAW_VISCOSITY_MODEL:
+    definition: Power Law model for a variable viscosity function that is limited by minimum and maximum values 
+    parent: CUBA.RHEOLOGY_MODEL
+    models: [CUBA.CONTINUUM]
+    CUBA.MAXIMUM_VISCOSITY: 
+      default: 1e-3
+    CUBA.MINIMUM_VISCOSITY:
+      default: 1E-5
```
tuopuu: use lower-case format of exponential numbers: 1e-5

simphony_metadata.yml
```diff
@@ -0,0 +1,326 @@
...
+      default: 1
+
+  CROSS_POWER_LAW_MODEL:
+    definition: Viscosity Cross power law model
+    parent: CUBA.RHEOLOGY_MODEL
+    models: [CUBA.CONTINUUM]
+    CUBA.INITIAL_VISCOSITY: 
+      default: 1e-3
+    CUBA.MAXIMUM_VISCOSITY: 
+      default: 1E-5   # nu_infiniy
```
tuopuu: lower-case format also here: 1e-5, and fix the comment line typo.

simphony_metadata.yml
```diff
@@ -0,0 +1,326 @@
...
+      default: 0.5
+
+  BIRD_CARREAU_MODEL:
+    definition: The Bird-Carreau model
+    parent: CUBA.RHEOLOGY_MODEL
+    models: [CUBA.CONTINUUM]
+    CUBA.INITIAL_VISCOSITY: 
+      default: 1e-3
+    CUBA.MAXIMUM_VISCOSITY: 
+      default: 1E-5
```
tuopuu: 1e-5

tuopuu:
@ahashibon, I saw just a few small typos, please see the line comments above. Otherwise it's good.

ahashibon:
@tuopuu @kitchoi you guys are doing a great job! thanks. is ok to merge now?

kitchoi:
Looks good to me! :+1:

tuopuu:
:+1:

PR #16, opened 2016-05-05
Cuds generator

roigcarlo:
@kitchoi Here is my work on the generator. I took a time to adapt it to some of the last wiki changes and clean up a little bit the code.

There are some aspects that need discussion. I will try to mark them in the code as the problems are pretty much similar as the ones you have in your branch before tomorrow morning.

Also pleas consider that all the documentation/testing part is still legacy code from the mater_relation generator. Expect lots of "mr" typos :/

Some important TODO's are:
- Shape attribute
- System data attribute keys exist but do not work since they relay on cuds_item's CUBA.DATA to be system managed, which is not ( is one of the points I want to discuss).


kitchoi:
@roigcarlo Do you mind if I attempt combining this branch with #8 ?  I can get back to you tomorrow for what I've got then we can move on from there.
Both branches give very similar results and there are bits here and there that can be put together.

roigcarlo:
@kitchoi Sure!, go ahead.

mehdisadeghi:
Being merged into #17

PR #17, opened 2016-05-10
Code generator combining PR 8 and PR 16

kitchoi:
Please try the following upon checking out the branch.
You should notice that there are some failures.  Fixing these failures require adding missing entries in cuba.yml and simphony_metadata.yml

```
$ pip install -r dev_requirements.txt
$ python setup.py develop
$ simphony-meta-generate meta_class simphony_metadata/yaml_files/simphony_metadata.yml simphony_metadata/scripts/tests/meta_class -O --test
$ simphony-meta-generate keywords simphony_metadata/yaml_files/cuba.yml simphony_metadata/scripts/tests/keywords.py
$ simphony-meta-generate cuba_enum simphony_metadata/yaml_files/cuba.yml simphony_metadata/yaml_files/simphony_metadata.yml simphony_metadata/scripts/tests/cuba.py
$ python -m unittest discover -v -p test*
```

TODO:
- [ ] populate the docstring for every attribute in a class
- [x] more tests on validation, (e.g. AttributeError on setting uuid etc.)

@roigcarlo please take a look.

@ahashibon could you try this out and fill in the missing cuba's please?

@mehdisadeghi I have a setup.py and .travis.yml here, but I don't have admin right on travis, could you set it up please?

Thanks.

PR #20, opened 2016-05-19
Fix missing CUBA and metadata inconsistencies

### LINKED ISSUE #18: Missing CUBA and corrections needed for the cuba.yml and simphony_metadata.yml
I had to make these changes to the yaml files in order to pass tests in #17 locally
cuba.yml:
https://gist.github.com/kitchoi/1815eeef45356a2eac0551845eea837c

simphony_metadata.yml
https://gist.github.com/kitchoi/aee20170f5d5690c95c153272a669a26

@ahashibon, @tuopuu  Can you have a look please?

kitchoi:
Fix #18 

Required by #17 

Please error travis fail for this PR as travis setup is included in #17

### DIFF
--- a/cuba.yml
+++ b/cuba.yml
@@ -9,7 +9,7 @@ CUBA_KEYS:
   UUID:
     definition: Universal unique id represented as a hex string size 32
     shape: [32]
-    type: string
+    type: uuid
 
   NAME:
     definition: Naming of high-level objects (e.g. solver models)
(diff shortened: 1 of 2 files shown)


kitchoi:
@tuopuu, @roigcarlo, @mehdisadeghi can we merge this?

roigcarlo:
Ok for my side

cuba.yml
```diff
@@ -9,7 +9,7 @@ CUBA_KEYS:
   UUID:
     definition: Universal unique id represented as a hex string size 32
     shape: [32]
-    type: string
+    type: uuid
```
mehdisadeghi: I think string is a better fit for metadata.

kitchoi: Okay, will do!

simphony_metadata.yml
```diff
@@ -16,7 +16,8 @@ CUDS_KEYS:
     parent:
     CUBA.UUID:
       scope: CUBA.SYSTEM
-    CUBA.DATA:
+    data:
```
mehdisadeghi, line 19: Even though I'm in favor of defining attributes in metadata level and just describe them with extra metatada (e.g. name, type, default, etc.), but the current consensus so far has been having fixed metadata attributes as lower case and dynamic ones as upper case with CUBA prefix. Therefore I think CUBA.DATA is the way to go for now.

simphony_metadata.yml
```diff
@@ -259,15 +259,14 @@ CUDS_KEYS:
...
-      shape: (2)
     CUBA.MATERIAL:
       shape: (2)
 
   DIRICHLET_BOUNDARY_CONDITION:
     definition: Dirichlet boundary condition
     parent: CUBA.CONDITION
     models: [CUBA.CONTINUUM]
-    CUBA.VARIABLES:
+    variables:
```
mehdisadeghi, line 269: Same as my previous comment. For plurals, we have to define something like CUBA.VARIABLE and describe it with extra metadata attributes such as shape. Anything with no `CUBA.` prefix implies a fixed metadata-level attribute. However, I think we should work on this whole thing and improve it.

simphony_metadata.yml
```diff
@@ -276,7 +275,7 @@ CUDS_KEYS:
     definition: Neumann boundary condition
     parent: CUBA.CONDITION
     models: [CUBA.CONTINUUM]
-    CUBA.VARIABLES:
+    variables:
```
mehdisadeghi, line 278: Same comment here.

simphony_metadata.yml
```diff
@@ -285,7 +284,7 @@ CUDS_KEYS:
     parent: CUBA.CONDITION
     definition: Empty boundary condition
     models: [CUBA.CONTINUUM]
-    CUBA.VARIABLES:
+    variables:
```
mehdisadeghi, line 287: The same comment here as well.

mehdisadeghi:
Thank you @kitchoi ! I just made a few minor comments.

I have one extra comment on defining default values. Currently we have two type of values, primary (defined in cuba.yml) and structures (simphony_metadata.yml). A primary value, e.g. integer, string, etc. is enough for primary CUBA but for structures I think we should put a yaml dump of all values of the corresponding structure for default. I'll try to do that in my PR for extra lammp keys.

simphony_metadata.yml
kitchoi, line 19: Isn't `data` a fixed metadata attributes?  CUBA.DATA is no primary type (but a container) that I think we would have problem defining it in cuba.yml (or any suggestion?)

simphony_metadata.yml
kitchoi, line 269: as far as I understand, 'variables' is the 'physical_variables' in https://publicwiki-01.fraunhofer.de/SimPhoNy-Project/index.php/SimPhoNy_Metadata_Schema#Converting_a_general_metadata_element_to_a_class
'variables' are also defined in other CUDS_KEYS (e.g. KS_DFT, CFD, ...).

simphony_metadata.yml
mehdisadeghi, line 19: Well, it's not. The value for `data` is undefined in metadata. As you mentioned, it is a structure and would be problematic to define. @ahashibon Any coment on this?

simphony_metadata.yml
mehdisadeghi, line 269: You are right, it is fixed, so ignore the comment.

kitchoi:
@mehdisadeghi Given that the CUBA.DATA vs data is the only question here, shall we create an issue about it and merge this one so that we can move on with generating the classes?

mehdisadeghi:
@kitchoi OK, lets address that issue in another thread!

PR #22, opened 2016-06-06
Add lammps keys

mehdisadeghi:
This PR adds some CUBA keys used in simphony-lammps-md project.

@ahashibon @TobiasRasp Please comment.

@kitchoi @tuopuu 
How do you find the structure of the BOX, specially the default field? I tried to dump the value of another structure as value for the default field.


simphony_metadata.yml
```diff
@@ -419,3 +419,28 @@ CUDS_KEYS:
...
+    parent: CUDS:COMPONENT
+    CUBA.POINT:
+      default: [0 0 0]
+
+  BOX:
+    definition: A simple hexahedron box object
+    parent: CUBA.BOUNDARY
+    CUBA.BASIS:
+      default:
+        CUBA.VECTOR: [[1, 0, 0], [0, 1, 0], [0, 0, 1]]
```
kitchoi:
```
CUBA.BASIS:
    default: [[1, 0, 0], [0, 1, 0], [0, 0, 1]]
```

should do.

simphony_metadata.yml
```diff
@@ -419,3 +419,28 @@ CUDS_KEYS:
   ELECTROSTATIC_MODEL:
     parent: CUBA.PHYSICS_EQUATION
     definition: Electrostatic model
+    physics_equation: [MOLECULAR_DYNAMICS]
+
+  BASIS:
```
kitchoi: Would it be possible to call it `BASIS_VECTORS`?  It would be more telling

simphony_metadata.yml
mehdisadeghi: In that case how can we recognize a default value for a primary CUBA from a default value for a structure with multiple CUBA values inside?

BTW, I think here I can remove the default all together, since it has the same value as in  BASIS itself.

simphony_metadata.yml
mehdisadeghi: It sounds better. I'll see.

simphony_metadata.yml
kitchoi: > a default value for a primary CUBA

By primary CUBA I understand that as attributes defined in cuba.yml.  There is no default value implemented there.

> a default value for a structure with multiple CUBA values inside

This is what 

```
BOX: 
     CUBA.BASIS:
         default : [[1, 0, 0], [0, 1, 0], [0, 0,1]]
```

means

simphony_metadata.yml
kitchoi: Or do you mean LATTICE_VECTORS, which is already defined in cuba.yml?

cuba.yml
```diff
@@ -545,3 +545,23 @@ CUBA_KEYS:
     definition: Classification of a fluid flow type
     shape: [1]
     type: integer
+
+  VECTOR:
+    definition: A vector
+    shape: [1]
+    type: double
```
kitchoi: CUBA.DIRECTION is defined in cuba.yml, is that what you want?

cuba.yml
```diff
@@ -545,3 +545,23 @@ CUBA_KEYS:
...
+
+  POINT:
+    definition: A point in a 3D space system
+    shape: [3]
+    type: double
```
kitchoi: CUBA.POSITION is defined in cuba.yml, is that what you want?

cuba.yml
```diff
@@ -545,3 +545,23 @@ CUBA_KEYS:
...
+
+  FACE:
+    definition: A face
+    shape: [1]
+    type: double
```
kitchoi: Face is defined in simphony.cuds.mesh and should have more info than just a numeric (double).
Could BOX use mesh or lattice internally and not have to define FACE here?

tuopuu:
I think @kitchoi is making a good point there in the line comments: We should try to use existing CUBA as much as it's possible, and avoid adding many similar keys in the list.

cuba.yml
ahashibon: > CUBA.DIRECTION is defined in cuba.yml, is that what you want?

direction is something else, it can be defines by a vector and origin system, or it may have other keys, such as up down, left right! we should ask those who defined it to elaborate it. 
a vector here is the mathematical basic definition. We need a better definition of direction and to see where it is used!

cuba.yml
ahashibon: > CUBA.POSITION is defined in cuba.yml, is that what you want?

a position is defined by a point. An entity can have a position, and this position is defined by a point in space. For example, a group of atoms may have a position that is defined as the centre of mass: group --> centre of mass --> position --> point.

cuba.yml
ahashibon: Good point! a face in a mesh is defined by the uuid of the nodes. The nodes can be used to define the basis vectors if additional information on the "order" of the nodes, i.e., how to map them to vectors is given (say first node is the origin, second defines the first vector, and so on). If we add this information to the basic definition of a cell, or alternatively define Box based on cell with additional attributes then we can assume that essentially a box is a special mesh! This should be added to the specifications!

simphony_metadata.yml
ahashibon: A LATTICE has an attribute VECTORS, propose to remove lattice_vectors to follow better the EAV model! 

a BASIS should have a vector (like defined already above) but of shape 3, we do not need additional lattice vectors or vectors!

simphony_metadata.yml
ahashibon: agree with @kitchoi

simphony_metadata.yml
tuopuu: > a BASIS should have a vector (like defined already above) but of shape 3, we do not need additional lattice vectors or vectors!

Agreed.

ahashibon:
> I think @kitchoi is making a good point there in the line comments: We should try to use existing CUBA as much as it's possible, and avoid adding many similar keys in the list. 

agree as long as the already existing CUBA make sense, and are used already deeply!

simphony_metadata.yml
mehdisadeghi: > By primary CUBA I understand that as attributes defined in cuba.yml

I mean the value that we assign to their CUBA entry (defined in a structure) not a `default` attribute inside cuby.yaml entries.

@ahashibon @kitchoi Still the problem remains: _how can we recognize a default value for a primary CUBA from a default value for a structure_. In the `CUBA.BASIS` example, how do we understand the the given value goes into `VECTOR` attribute of the `BOX` and not another one? (a structure could have multiple attributes).

kitchoi: @mehdisadeghi I think there is a limit on what you can do with just yaml files.  In this case since BASIS is defined with a default value of VECTOR, in BOX you may try this

```
BOX:
    CUBA.BASIS:
        default : CUBA.BASIS
```

See

```
CFD:
   CUBA.COMPRESSIBILITY_MODEL:
       default: CUBA.INCOMPRESSIBLE_FLUID_MODEL
```

cuba.yml
mehdisadeghi: @ahashibon @kitchoi which changes do you propose for FACE?

simphony_metadata.yml
mehdisadeghi: @kitchoi I will change this as you suggested, we can come back to it later, when necessary.

mehdisadeghi:
@kitchoi @ahashibon This PR is finally passes the test. I had a problem with `default` value for `BASIS.VECTOR` and I had to put `shape = (3, 3)` to get rid of it.

kitchoi:
@mehdisadeghi Can you update your local master branch, merge master to this branch and then push your changes here please?

mehdisadeghi:
@kitchoi There are no new changes in master to be merged with this branch. Is there anything specific to be merged?

kitchoi:
@mehdisadeghi I just thought that might be why travis ci test for PR was failing...maybe not then.

mehdisadeghi:
@kitchoi Would you mind checking the error message more in detail? I just tried running tests in CI container and realized that in the case of the failing test, most probably, the mock patch is not applied.

kitchoi:
@mehdisadeghi Ok, I am looking into it.

yaml_files/simphony_metadata.yml
```diff
@@ -418,5 +418,36 @@ CUDS_KEYS:
...
-    parent: CUBA.PHYSICS_EQUATION
     definition: Electrostatic model
+    parent: CUBA.PHYSICS_EQUATION
+    models: [CUBA.CFD, CUBA.MOLECULAR_DYNAMICS, CUBA.DEM]
+
+  BASIS:
+    definition: Space basis vectors
+    parent: CUBA.CUDS_COMPONENT
+    CUBA.VECTOR:
+      shape: (3, 3)
```
ahashibon, line 429: This should be shape: 3 
the Basis is a list of three vectors (which in turn are of shape 3).

yaml_files/simphony_metadata.yml
```diff
@@ -418,5 +418,36 @@ CUDS_KEYS:
...
+    CUBA.VECTOR:
+      default: [0, 0, 0]
+      scope: CUBA.USER
+    CUBA.FACE:
+      default: 0
+      scope: CUBA.USER
+    CUBA.BASIS:
+      default: CUBA.BASIS
+      scope: CUBA.USER
+    CUBA.ORIGIN:
```
ahashibon: Do we need origin here? if we define it then we would need to define a global coordinate system, Since there is no need for it now, how about removing it!

yaml_files/simphony_metadata.yml
kitchoi, line 429: The default value is `[[1, 0, 0], [0, 1, 0], [0, 0, 1]]` which is of shape (3, 3).
Or do we only check the first dimension?

yaml_files/simphony_metadata.yml
```diff
@@ -418,5 +418,36 @@ CUDS_KEYS:
...
+      default: [[1, 0, 0], [0, 1, 0], [0, 0, 1]]
+
+  ORIGIN:
+    definition: The origin of a space system
+    parent: CUBA.CUDS_COMPONENT
+    CUBA.POINT:
+      default: [0, 0, 0]
+      scope: CUBA.USER
+
+  BOX:
```
ahashibon, line 439: Box should have currently only the vector (of shape 3). Although we can define it as a simple cell, this should be avoided as the parent of BOX should be a Boundary! a cell is a low level computational object, while the metadata should reflect the logic of the scientific end user problem.

kitchoi:
@mehdisadeghi I ran the tests with the travis docker image too and I could not reproduce the errors.  Could you reproduce it?  If so, you may be at a better position than me for hunting down the source of errors.

kitchoi:
Turns out test failure is caused by the different order of tests run in travis-ci/pr and travis-ci/push.
The errors are reproducible by running this:

```

$ python -m unittest discover -v -p test_v*
test_check_shape_simple (simphony_metadata.scripts.tests.test_validation.TestValidation) ... ok
test_check_shape_special (simphony_metadata.scripts.tests.test_validation.TestValidation)
Test for value=1 and value=[1] to be both valid for shape=(1) ... ok
test_decode_shape (simphony_metadata.scripts.tests.test_validation.TestValidation) ... ok
test_error_check_shape (simphony_metadata.scripts.tests.test_validation.TestValidation)
Test for ValueError for invalid shapes ... ok
test_error_validate_cuba_keyword (simphony_metadata.scripts.tests.test_validation.TestValidation)
Test for TypeError for invalid CUBA keyword value ... ERROR
test_validate_cuba_keyword (simphony_metadata.scripts.tests.test_validation.TestValidation)
Test for valid cases for CUBA keyword values ... simphony_metadata/scripts/tests/meta_class/validation.py:166: UserWarning: Value is a string, its shape is not validated. Please fix the cuba.yml shape syntax.
  warnings.warn('Value is a string, its shape is not validated. '
ok

======================================================================
ERROR: test_error_validate_cuba_keyword (simphony_metadata.scripts.tests.test_validation.TestValidation)
Test for TypeError for invalid CUBA keyword value
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "simphony_metadata/scripts/tests/test_validation.py", line 99, in test_error_validate_cuba_keyword
    type=KEYWORDS[cuba_name.upper()].dtype,
KeyError: 'CUBA.MATERIAL'

----------------------------------------------------------------------
Ran 6 tests in 0.004s

FAILED (errors=1)
```

The patch magic in test_validation has not worked and probably only had worked relying on imports from another test module.

mehdisadeghi:
@kitchoi Your patch fixed the error! However I am adding more commits to this PR and then it is ready to be merged.

yaml_files/simphony_metadata.yml
mehdisadeghi, line 429: @kitchoi I think we have to add the `shape` defined in this file to the `shape` defined in the `cuba.yaml` file and then validate the value. Therefore, a shape of `(3)` for `BASIS.CUBA.VECTOR` would imply a _real_ shape of `(3, 3)` for it.

mehdisadeghi, line 429: At the moment I have to leave shape as is, until we reach an agreement, move on and patch the generator.

PR #23, opened 2016-06-09
Patch CUBA before import.

mehdisadeghi:
@kitchoi This patch fixes the import error which I described in #17


kitchoi:
:+1: Thank you!

PR #26, opened 2016-06-16
Attributes not begining with CUBA are system-managed and read-only

kitchoi:
Tests would fail, awaiting @mehdisadeghi's PR on changing some of the attributes back to fixed attributes.

### DIFF
--- a/simphony_metadata/scripts/generate.py
+++ b/simphony_metadata/scripts/generate.py
@@ -25,12 +25,9 @@ IMPORT_PATHS = {
     'validation': 'from . import validation'
     }
 
-# These are keys that are read-only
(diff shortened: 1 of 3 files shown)


mehdisadeghi:
Thanks for the quick PR. 

Regarding the title, anything with no `CUBA.` in the beginning, is basically fixed metadata information and no variable at all.

kitchoi:
@mehdisadeghi Thanks.  I corrected the title.

kitchoi:
@mehdisadeghi Done.  Ready for review.

mehdisadeghi:
@kitchoi It looks good to me.

PR #28, opened 2016-06-17
Add NCad CUBA keywords. Deleted wrong ones (LATTICE_VECTORS and SYM…

SGGgarcia:
- Added NCad CUBA keywords:
- Deleted wrong ones (LATTICE_VECTORS and SYMMETRY_LATTICE_VECTORS)


yaml_files/cuba.yml
```diff
@@ -565,3 +555,59 @@ CUBA_KEYS:
     definition: Thermodynamic ensemble
     shape: [1]
     type: string
+
+  LATTICE_UC_ABC:
```
ahashibon, line 559: LATTICE_VECTORS is more "text book" standard than LATTICE_UC_ABC, please revert to using the more standard terminology.

yaml_files/cuba.yml
```diff
@@ -565,3 +555,59 @@ CUBA_KEYS:
...
+    definition: Lattice translation vectors lengths (a, b, c) of an atomic unit cell
+    shape: [3]
+    type: double
+
+  LATTICE_UC_ANGLES:
```
ahashibon, line 564: Please remove UC, instead its LATTICE_ANGLES or expand UC to UNIT_CELL. Please note that this does not give any information of whether this is a primitive unit cell or not! Alternatively we can think of a CRYSTAL_ LATTICE as a data structure in itself, that has a unit cell that has vectors and angles and symmetry etc.

yaml_files/cuba.yml
```diff
@@ -565,3 +555,59 @@ CUBA_KEYS:
...
+  CRYSTAL_ORIENTATION_1:
+    definition: Crystallographic orientation defined by Miller Index of the reference surface
+    shape: [3, 3]
+    type: double
+
+  CRYSTAL_ORIENTATION_2:
+    definition: Crystallographic orientation defined by Miller Index of the non-reference surface
+    shape: [3, 3]
+    type: double
+
```
ahashibon, line 583: Please remove _1 and _2, there is only one orientation (do we really need the extra crystal_orientation_ prefix ?)
A crystal object is what we need to add that has an orientation which has a vector? 
why shape is 3x3?

yaml_files/cuba.yml
```diff
@@ -565,3 +555,59 @@ CUBA_KEYS:
...
+  STL_SCALING:
+    definition: Scaling parameter for components with STL shape
+    shape: [1]
+    type: double
+
+  STL_PADDING:
+    definition: Padding width in -X, +X, -Y, +Y, -Z and +Z directions for components with STL shape
+    shape: [6]
+    type: double
+
```
ahashibon, line 613: This needs to be refined by defining an "STL" object with attributes that can be a FILE etc. Please define it in such a way that FILE is general (i.e., can be used by other objects).

SGGgarcia:
Thanks for the comments, @ahashibon , I'll try to answer everything and let @gromansg answer with additional information in case we need it

yaml_files/cuba.yml
SGGgarcia, line 613: There is a cuba called `FILE_STL` in this PR that defines a path for a STL file. The STL_MODE, STL_SCALING and STL_PADDING new CUBAS are nCad attributes, in concrete they are attributes of the nCad components that will have a STL shapes (this means, the shape they will have is defined by the triangles contained in the .stl file that make a 3D surface). They are not related to the .stl file indeed, they are just like any other attribute

yaml_files/cuba.yml
SGGgarcia, line 583: Indeed you are right, there is only one crystallographic orientation. These two are different orientations: one is the crystallographic orientations as we know its (how the replicated cells are going to be orientated along the component shape) and the other is the orientation of the full shape in the global 3D axis. The shape is 3x3 as these orientations define two new 3D references to orientate the building process of the component and the position of it
What I agree is that maybe the names are not really that clear, we will think about this

yaml_files/cuba.yml
SGGgarcia, line 559: We changed this as they are the `a b c` parameters of the cell and not the full translation vectors of the cell

yaml_files/cuba.yml
SGGgarcia, line 564: I'll look into this aswell

yaml_files/cuba.yml
gromansg: Lattice vectors refers to the three 3D vectors that define the unit cell. Here we are not requesting these vectors but the lattice parameters (a,b,c and alpha, beta, gamma) that are the most common way to define the Unit Cell (even more common that the vectors), in particular in atomics builders. In our concern, it is much more easier to ask to the user for a set of 3 length parameters (a,b,c) and a set of three angle parameters (alpha, beta, gamma) than ask for the six parameters at the same time. If we use "Lattice parameters" we have to ask for 6 parameters. That is the reason we have add a "surname" to these two CUDAs. Anyway, name of the CUDA could be discussed, but we cannot change to Lattice Vectors. Besides, it is not so standard as Lattice Parameters

yaml_files/cuba.yml
gromansg: See previous comment

yaml_files/cuba.yml
gromansg: Yes, you are right, in principle there is only one crystallographic orientation defined by Index Miller (h,k,l). However, this description leave 1 degree freedom (the rotation around the axe defined by the Index Miller direction). This is usually not important because the scientist usually works in theoretical framework, with a fixed reference system. However, When you translate this to the "real" 3D space, n in determination appears. This indeterminacy must be defined by user with another vector that we called Crystal orientation 2. This value can be understand as a fake index Miller of one of the perpendicular direction to the Index Miller. Of course, not all the values are permitted for this   fake index miller, because it is conditioned by the real Index Miller and the crystal structure. We check the input given by the user, if it is correct we used, if not we used the closest correct orientation vector. 
In nCAD GUI we also give the opportunity to give an angle to define this indeterminacy. The problem is that the angle is not completely defined because it is not defined a 0 for this 0 must to arbitrary chosen (even if you use a criterion you are setting it arbitrarily). In the GUI it is easier to use it because it is graphically guided, but in an API it is a little bit confusing.

gromansg: It is not a 3x3, it is a bug. It is 2x3, two components of three elements each on. First component is the Miller Index (and the fake Index Miller in orientation_2) and the second one is the real space direction that you want your Miller direction matchs.

yaml_files/cuba.yml
gromansg: As Goyo has remarked, all these CUBAs are nCAD attributes. They are used to poisoned an STL drawn in the atomic reference system:
- STL_MODE (0/1): 0 means that nCAD will use the STL drawn directly and will fill with atoms the shape contained by the triangles of the STL. 1 means that nCAD will build another STL shape that it is a cubic shape where the original STL is removed. A clear example is a complex assembly of channels or a porous like material with this hollow channels inside. 
- STL_Scaling: apply an scale factor to the STL. Hence, a user can use a STL created by traditional CAD (where you are using mm, microns, meters, ...) and imported it to nCAD in nanometric scale. 
- STL_Padding: It is used to move and set your 0 in the Unit cell and therefore translate the atomic structure distances minor than the UC

yaml_files/cuba.yml
gromansg: About your suggestion, we are not able to change this now, because it means deep changes in the wrapper/nCAD API. This CUBAs where defined long time ago (in the first moths of the projects) and implementation work with them.
Of course we can discuss if the name is the best one or it must be modified
In the next implementations we already implementing with your suggestions.

mehdisadeghi:
Please consider the latest changes to metadata. We have introduced lattice and mesh related changes.

mehdisadeghi:
Since there are conflicts and there has been no activity on this PR for a long time, I'm going to close this. Feel free to reopen and continue it though.

PR #37, opened 2016-07-14
Generator screens defaults for MutableSequence

### LINKED ISSUE #36: Some mutable defaults are not replaced with None in the generator
The generator has taken care of `default = []` and replaced it with `None` in the `__init__`, but not for defaults such as `[0., 0., 0.]`, so you end up with `def __init__(something=[0., 0., 0.])`.

For the current generated classes, there is however no sharing of mutable objects among instances because these default values are first copied and validated before being saved to the data container.  However, it is considered a common practice not to have mutables as arguments' defaults.

kitchoi:
Fix #36

### DIFF
--- a/simphony_metadata/scripts/generate.py
+++ b/simphony_metadata/scripts/generate.py
@@ -5,7 +5,7 @@ import re
 import shutil
 import tempfile
 import warnings
-from collections import OrderedDict
+from collections import OrderedDict, MutableSequence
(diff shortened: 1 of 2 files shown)


kitchoi:
@mehdisadeghi Please review.  Thanks.

mehdisadeghi:
It looks fine to me.

PR #42, opened 2016-08-11
Add legacy yaml files and generators.

mehdisadeghi:
For reference, here are the former generator and yaml files which are of no use anymore.

@tuopuu @kitchoi @roigcarlo @SGGgarcia Please see the changes.


SGGgarcia:
:+1:

Issue #39, opened 2016-07-21
Consider moving the code generator back to simphony-common


kitchoi:

stefanoborini:
@mehdisadeghi Where is the code generator supposed to be? I suspect it should be in simphony-metadata, so that when the metadata files change in a non-backward compatible way, the generator follow in the change. 

Right now we have two copies of the generator, one in simphony-metadata, and another in simphony-common. One must go. Which one?

mehdisadeghi:
@stefanoborini previously with @kitchoi we discussed this and decided to move it back to simphony-common. However, the problem of running tests upon changes to metadata remains unaddressed.

stefanoborini:
@mehdisadeghi ok which means that the copy currently in simphony-metadata is not supposed to be there. For the tests, when the metadata automatic build will be merged, we will always run against the most recent commit of the metadata.

mehdisadeghi:
> which means that the copy currently in simphony-metadata is not supposed to be there

There is a PR #49 for that. However, we still need to add a YAML and schema validator to this repo.

mehdisadeghi:
Done in #49

PR #46, opened 2016-10-06
Problem: tests fail because of patching problem.

mehdisadeghi:
Solution: patch the the missing spots.

Upon import the class definition loads once and therefore the class variables remain unchanged even after patching. This PR adds them to the patch list.


kitchoi:
What's the status of #39?  It would be so much easier if the code generator for simphony-common sits inside simphony-common.

mehdisadeghi:
> What's the status of #39?

Waiting in my TODO list.

PR #47, opened 2016-10-11
PR: Changing CFG.GRAVITY_MODEL to have a scope

tuopuu:
When instantiating a Cfd class a gravity model is required to be given at the initialization. This prevents easy handling and instantiation of CUDSComponents loaded from serialized CUDS/yaml files. To work around this problem, we should give a scope parameter to Cfd.gravity_model so that the code generator allows instantiating the Cfd class with 

```
gravity_model = None.
```

It would be similar how it's done in the case of MaterialRelation.


tuopuu:
@mehdisadeghi , could you please check that the generated code actually changes? (I can't make it work with my installation)

yaml_files/simphony_metadata.yml
```diff
@@ -170,6 +170,7 @@ CUDS_KEYS:
     CUBA.RHEOLOGY_MODEL:
       default: CUBA.NEWTONIAN_FLUID_MODEL
     CUBA.GRAVITY_MODEL:
+      scope: CUBA.USER
```
mehdisadeghi, line 173: @tuopuu If we provide a value for `default` here, it will be possible to instantiate `Cfd` without passing any arguments.

mehdisadeghi: The quick solution is putting `default:`. It will pass `None` as default value.

PR #48, opened 2016-10-11
Problem: shape of electrostatic_field should be 3

mehdisadeghi:
Solution: set it to 3

@tuopuu please review this change.


tuopuu:
This works. Thanks @mehdisadeghi

PR #49, opened 2016-10-17
Problem: python class generator belongs to simphony-common repository

mehdisadeghi:
Solution: remove it from this repository (and move it to simphony-common)

Along the way I have remove the tests from this repo. I can move them back but then we have to use simphony-common's tests to check this repo. Leave a common if you think it is necessary.


stefanoborini:
@mehdisadeghi can I complete this one?

mehdisadeghi:
@stefanoborini sure you can :)

PR #50, opened 2016-10-25
Problem: domain entities are missing from the metadata

mehdisadeghi:
Solution: add them all.

This PR defines `PARTICLE`, `PARTICLES`, `LATTICE`, `MESH` and many related entities.


yaml_files/cuba.yml
```diff
@@ -567,10 +567,10 @@ CUBA_KEYS:
     shape: [3]
     type: double
 
-  POINT:
-    definition: A point in a 3D space system
+  INDEX:
+    definition: An integer vector in 3D geometirc space
```
stefanoborini: geometric

PR #52, opened 2016-10-26
Problem: some CUBAs lack shape attribute

mehdisadeghi:
Solution: change [] to [1] to have at least one dimention


stefanoborini:
This supposedly fixes the problems I had in simphony/simphony-common#329

PR #53, opened 2016-10-27
Problem: BRAVAIS_LATTICE entry has ill-formatted PRIMITIVE_CELL default value

mehdisadeghi:
Solution: remove the default value 

`BRAVAIS_LATTICE` CUBA entry has an array as its default value for `PRIMITIVE_CELL` attribute, however, we have so far no way to assign a default value for non-primitive CUBAs. Therefore this causes problems and I had to remove it and rely on the default value defined in `PRIMITIVE_CELL` itself.


mehdisadeghi:
@ahashibon please see this change.

@stefanoborini please review this

mehdisadeghi:
This change is already applied to simphony/simphony-common#329

PR #56, opened 2016-11-04
Removed data as it's now internalized information

stefanoborini:
Refers to PR simphony-common/346


mehdisadeghi:
@ahashibon your input is necessary.

mehdisadeghi:
At the moment we want `data` to be part of the metadata. We have to deal with this problem differently.

Related PR: https://github.com/simphony/simphony-common/pull/346

ahashibon:
Please do not remove it from the metadata, add temporary check in the code instead. This is a vital element.

PR #57, opened 2016-11-04
Fix floating points values representation

stefanoborini:
Fixes the representation of floating points, which with the given format is interpreted as a string by the yaml parser.


stefanoborini:
Incorrect basing. Closing and reopening

PR #62, opened 2016-11-10
Data as system variable

stefanoborini:
Set data as system variable to remove it from the initialization requirements


stefanoborini:
This PR currently contains the repotag to build the current master on sim-common. sim-common will need a PR to bind against the merge commit on master here, as soon as it's merged.

PR #67, opened 2016-11-23
Added format docs

stefanoborini:
Describing an official specification of the yaml file format.


ahashibon:
can be merged!

stefanoborini:
@ahashibon not yet. For two reasons:
- I think that what we have in the wiki should be here. The wiki is not contextual information, it's awkward to use, and doesn't follow up with the changes in code. 
- I'm still not done.

PR #73, opened 2016-12-08
supplement cfd keyword and metadata

khiltunen:
@ahashibon could you please review these additions on cfd keywords and metadata

### DIFF
--- a/yaml_files/simphony_metadata.yml
+++ b/yaml_files/simphony_metadata.yml
@@ -497,7 +497,7 @@ CUDS_KEYS:
       default: 1.0
     CUBA.POWER_LAW_INDEX:
       default: 1.0
-  
+
   INLET_OUTLET:
     definition: Inlet outlet boundary condition (outlet condition is zero gradient and inlet given variable value)
     parent: CUBA.CONDITION
--- a/yaml_files/simphony_metadata.yml
+++ b/yaml_files/simphony_metadata.yml
@@ -486,9 +486,7 @@ CUDS_KEYS:
 
   SHEAR_STRESS_POWER_LAW_SLIP_VELOCITY:
     parent: CUBA.CONDITION
-    definition: Shear stress power law dependant slip velocity boundary condition. Nonlinear boundary condition for wall tangential velocity of the form
-v_s = CUBA.LINEAR_CONSTANT * S^CUBA.POWER_LAW_INDEX
-where v_s is the slip velocity (tangential velocity on the wall) and S is the wall shear stress
+    definition: Shear stress power law dependant slip velocity boundary condition. Nonlinear boundary condition for wall tangential velocity of the form v_s = CUBA.LINEAR_CONSTANT * S^CUBA.POWER_LAW_INDEX where v_s is the slip velocity (tangential velocity on the wall) and S is the wall shear stress
     models: [CUBA.CONTINUUM]
     CUBA.VARIABLE:
       shape: (:)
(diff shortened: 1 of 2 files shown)


yaml_files/simphony_metadata.yml
```diff
@@ -284,6 +298,40 @@ CUDS_KEYS:
     parent: CUBA.PHYSICS_EQUATION
     definition: Multiphase model
 
+  RELATIVE_VELOCITY_MODEL:
+    parent: CUBA.PHYSICS_EQUATION
+    definition: Relative velocity model to use in mixture model
+    models: [CUBA.CONTINUUM]
+
+  MESO_SCOPIC_RELATIVE_VELOCITY_MODEL:
```
stefanoborini: Typo? should be MESOSCOPIC

yaml_files/simphony_metadata.yml
```diff
@@ -284,6 +298,40 @@ CUDS_KEYS:
...
+    parent: CUBA.PHYSICS_EQUATION
+    definition: Stress model to use in mixture model
+    models: [CUBA.CONTINUUM]
+
+  STANDARD_STRESS_MODEL:
+    parent: CUBA.STRESS_MODEL
+    definition: Standard stress model (to use in mixture model)
+    models: [CUBA.CONTINUUM]
+
+  MESO_SCOPIC_STRESS_MODEL:
```
stefanoborini: Same

yaml_files/simphony_metadata.yml
```diff
@@ -436,6 +484,50 @@ CUDS_KEYS:
     definition: Periodic boundary condition (PBC)
     models: [CUBA.ELECTRONIC, CUBA.ATOMISTIC, CUBA.MESOSCOPIC, CUBA.CONTINUUM]
 
+  SHEAR_STRESS_POWER_LAW_SLIP_VELOCITY:
+    parent: CUBA.CONDITION
+    definition: Shear stress power law dependant slip velocity boundary condition
+    models: [CUBA.CONTINUUM]
+    CUBA.VARIABLE:
```
stefanoborini, line 491: I am a bit baffled by the generic nature of this entity... Maybe I don't get its point.

khiltunen: It's an nonlinear boundary condition for wall tangential velocity of the form
v_s = k S^n
where k and n are parameters and v_s is the slip velocity (tangential velocity on the wall) and S is the wall shear stress

stefanoborini: What I am saying is that the name "VARIABLE" is too generic to be understood without more information (as from your comment).

ahashibon: Each condition can be applied on one or more field variables (in general on the whole domain or specifically on boundaries). So each condition has a CUBA.VARIABLE property which is a list of variables that the condition applies to. These variables themselves should be defined in the CUBA set (currently at least in cuba.yml). We could have called it AFFECTED_VARIABLE, but this is unnecessary as it is clear from the context to intended users and will just increase the amount of CUBA without real need. However, since in this particular condition, the velocity is the only affected variable, one could set it on the metadata level by using lower case: variable: CUBA.VELOCITY (however, we should be clear whether this needs to apply to one particular phase?)

stefanoborini:
Don't worry about the conflicts, just ask me if it's not intuitive. Let me know when you are ready.

stefanoborini:
@khiltunen What is the relationship between this PR and the one we just merged? Is it a duplicate?

khiltunen:
We can forget this PR. These are not duplicate but the same data is included in the merged PR.

PR #93, opened 2016-12-29
Basic implementation of non-empty models [EXP]

stefanoborini:
Enforces the presence of all computational models in the base item.


stefanoborini:
Maybe even at the CUDS_ITEM level?

ahashibon (commented):
this will create a funny cycle for the COMPUTATIONAL_MODEL items, since they define the models themselves, and are cuds components.

stefanoborini:
@ahashibon it's not a cycle. it's a reimplementation.

PR #96, opened 2016-12-30
Format: Removed default values. Converted shape to length for strings.

### LINKED ISSUE #60: yaml file "shape" specification is ambiguous for strings
shape is used inconsistently in the metadata. 

When used with a numerical type, it indicates the number of entries, in this case, a linearized 3x3 matrix.

```
410   STRESS_TENSOR:
411     definition: Stress tensor
412     shape: [9]
413     type: double
```

In the case of string, it seems to represent the length of this string

```
 10   UUID:
 11     definition: Universal unique id represented as a hex string size 32
 12     shape: [32]
 13     type: string
```

Which is ambiguous and inconsistent with the precedent case. I would expect an array of 32 strings in this case.

stefanoborini:
Fixes #60

### DIFF
--- a/yaml_files/cuba.yml
+++ b/yaml_files/cuba.yml
@@ -9,12 +9,12 @@ CUBA_KEYS:
 
   UUID:
     definition: Universal unique id represented as a hex string size 32
-    length: [32]
+    length: 32
     type: string
 
   NAME:
     definition: Naming of high-level objects (e.g. solver models)
-    length: [20]
+    length: 20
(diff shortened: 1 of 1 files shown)


ahashibon:
reviewed changes, all good to go.

stefanoborini:
@ahashibon We need to remember that this change breaks the generator. I'll fix that as part of getting 0.5.0 (which is what's priority as we get into new year)

PR #97, opened 2016-12-30
Unified the cuba.yml and simphony_metadata.yml files. [EXP]

### LINKED ISSUE #92: Merge cuba.yml and simphony_metadata.yml
It makes no sense that we have two files, especially when the information can be easily put in the same file easily. This will hugely simplify the validator and the parser for the simphony-metaeditor.

stefanoborini:
Fixes #92

The generator needs to be adjusted to handle the new layout.
The metaeditor has the same need.
We also need to update the format specs.


stefanoborini:
Opened for discussion. I think it's a good idea, but it implies a lot of changes.

stefanoborini:
I recommend this unification to be done after 0.5.0. The current generator works as is, and we keep it as valid for now. As soon as we release 0.5.0, in an effort to streamline the file, we can apply this change

stefanoborini:
Closing. We are moving away from the YAML format

PR #102, opened 2017-01-05
Simplified and specified the meaning of the shape

stefanoborini:
Defined the meaning of the default, and using [] when appropriate


ahashibon:
If not mistaken we had a discussion about mixing () and [] for shape @kitchoi @mehdisadeghi any comment from your side? to me it looks ok to mix.

stefanoborini:
@ahashibon I am not aware of the discussion, but i suspect it's because we would not otherwise be able to specify open ended dimensions. [] is parsed as a list, and [:] is not a valid list. Using () gets parsed as a string (not a tuple, luckily), so we can pass the syntax we want, and parse it ourselves with the generator.

ahashibon:
Exactly! That was the discussion.

PR #121, opened 2017-03-13
Add entries for PDE conditions

ahashibon:
add missing value to conditions: a PDE condition needs a value, one variable, and an optional material (phase) it acts on. If no material is specified, it is assume that the condition applies universally to all materials.


ahashibon:
@khiltunen @kemattil please check these changes mainly for PDE boundary conditions.

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +444,34 @@ CUDS_KEYS:
...
   ######################################################
-  DIRICHLET:
-    definition: Dirichlet boundary condition, specify the value the solutions takes on the boundary of the domain
+  VALUE_CONDITION:
+    definition: Boundary condition encountered in the solution of partial differential equations
     parent: CUBA.CONDITION
     models: [CUBA.CONTINUUM]
     CUBA.VARIABLE:
-      shape: (:)
+      shape: [1]
```
stefanoborini: Not needed. it's the default

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +444,34 @@ CUDS_KEYS:
...
       default: []
-    CUBA.MATERIAL:
+    CUBA.VALUE:
       shape: (:)
+      default: 0.0
```
stefanoborini: Shape is arbitrary means this must be a list.

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +444,34 @@ CUDS_KEYS:
...
+    CUBA.MATERIAL:
+      shape: [1]
```
stefanoborini: Not needed.

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +444,34 @@ CUDS_KEYS:
...
       default: []
```
stefanoborini: If the shape is 1, this is no longer a list.

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +444,34 @@ CUDS_KEYS:
...
 
+
+
+  DIRICHLET:
```
stefanoborini: This change breaks backward compatibility.

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +444,34 @@ CUDS_KEYS:
...
     definition: Empty boundary condition
-    models: [CUBA.CONTINUUM]
-    CUBA.VARIABLE:
-      shape: (:)
-      default: []
-    CUBA.MATERIAL:
-      shape: (:)
-      default: []
+    parent: CUBA.VALUE_CONDITION
+    value: CUBA.NONE
```
stefanoborini: We never used this key, and it's not in the usual fixed variables

yaml_files/simphony_metadata.yml
```diff
@@ -503,32 +496,21 @@ CUDS_KEYS:
...
-    parent: CUBA.CONDITION
+    parent: CUBA.VALUE_CONDITION
     definition: Volume fraction wall boundary condition with specified contact angle
     models: [CUBA.CONTINUUM]
-    CUBA.VARIABLE:
-      shape: (:)
-      default: []
-    CUBA.CONTACT_ANGLE:
-      default: 45.0
+    variable: CUBA.VOLUME_FRACTION
```
stefanoborini: We have no `variable`. We have `variables`, and it must be a list.

stefanoborini (changes requested):

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +444,34 @@ CUDS_KEYS:
...
-  DIRICHLET:
-    definition: Dirichlet boundary condition, specify the value the solutions takes on the boundary of the domain
+  VALUE_CONDITION:
+    definition: Boundary condition encountered in the solution of partial differential equations
     parent: CUBA.CONDITION
     models: [CUBA.CONTINUUM]
     CUBA.VARIABLE:
-      shape: (:)
+      shape: [1]
       default: []
```
stefanoborini: This must not be a list, if the shape is 1

stefanoborini (changes requested):

yaml_files/simphony_metadata.yml
ahashibon: agree, but can we leave it for now to make sure the change is clear!

stefanoborini: This information goes in the changelog, not in the code.

yaml_files/simphony_metadata.yml
ahashibon: what value will it have if we do not set a default?
or can't we set the default value to any element in the list?

yaml_files/simphony_metadata.yml
stefanoborini: Besides, this change of course also ruins backward compatibility.

ahashibon: it has not been used so far though!

yaml_files/simphony_metadata.yml
stefanoborini: Point is that it should be `[0.0]`

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +444,34 @@ CUDS_KEYS:
...
 
+
+    
+  DIRICHLET:
+    definition: Dirichlet boundary condition to specify the value the solutions takes on the boundary of the domain.
+    parent: CUBA.VALUE_CONDITION
+
   NEUMANN:
```
khiltunen: If we are not using Dirichlet/Neumann BC names, i think thaat equally also Neumann should be changes to for example GRADIENT_CONDITION

yaml_files/simphony_metadata.yml
ahashibon: it merely means that there is no default. CUBA.NONE is defined in cuba.yaml. I assume that so far an empty default is equivalent to NONE?

yaml_files/simphony_metadata.yml
ahashibon: by specifications: anything in lowercase is fixed in the metadata level, so variable is equivalent to CUBA.VARIABLE. 
we should probably at some point formalise the plural (s) use! If now variables is equivalent to. 
CUBA.VARIABLE: 
  shape: (:) 
then we can use variables.

yaml_files/simphony_metadata.yml
stefanoborini: That's now how defaults are supposed to work. You specify the default in the receiving end, not the cuds declaration

yaml_files/simphony_metadata.yml
stefanoborini: > variable is equivalent to CUBA.VARIABLE.

It is not. not according to the current specs. Otherwise, we need to add `models` as well as CUBA.MODELS, and since we are there, also parent. 

Besides, using the case to define this is very deceiving and unheard of. The point is that if you want a variable property to be readonly, we should have a way of doing so.

stefanoborini: If you want variable properties to be readonly, then the distinction between fixed and variable properties is broken and we should have cuba vs. non-cuba properties. In that case, I need to be able to specify a cuba property as readonly, something I cannot currently do. This will mean changing the specs, changing the file and changing the parser.

stefanoborini: Specs are here

https://raw.githubusercontent.com/simphony/simphony-metadata/master/docs/format.md

Anything else is too old/informally specified to still be relevant.

stefanoborini:
After discussion offline, this PR can only be accepted with a considerable change in our specifications, parser, and overall vision. If we do so, we must do as indicated in issue #122

ahashibon:
this is WIP. Just changed variables to have system scope for slip velocity. @stefanoborini would this work with current scope of the specs? if not will change back to "normal" CUBA.VARIABLE behaviour for now. 

We are conducting small ontology analysis of boundary conditions, will update soon.

yaml_files/simphony_metadata.yml
```diff
@@ -503,32 +494,22 @@ CUDS_KEYS:
...
 
   SLIP_VELOCITY:
-    parent: CUBA.CONDITION
+    parent: CUBA.VALUE_CONDITION
     definition: Wall free slip velocity boundary condition
     models: [CUBA.CONTINUUM]
     CUBA.VARIABLE:
-      shape: (:)
-      default: []
+      default: CUBA.VELOCITY
```
stefanoborini: CUBA.VELOCITY is not a CUBA.VARIABLE, what are you trying to achieve here?

yaml_files/simphony_metadata.yml
```diff
@@ -488,12 +482,9 @@ CUDS_KEYS:
     models: [CUBA.ELECTRONIC, CUBA.ATOMISTIC, CUBA.MESOSCOPIC, CUBA.CONTINUUM]
 
   SHEAR_STRESS_POWER_LAW_SLIP_VELOCITY:
-    parent: CUBA.CONDITION
+    parent: CUBA.VALUE_CONDITION
```
stefanoborini: This type is not declared anywhere

stefanoborini (changes requested):

ahashibon:
CUBA.VELOCITY is a CUBA.VARIABLE, for now what we have in cuba,yaml as "variables", and possibly other fields if defined even in simphony-metadata should be allowed for CUBA.VARIABLE. @khiltunen please comment. 

Each condition works on a specific variable/field, and this has to be specified in the metadata. These, i.e., the variables and fields are part of the entities specified in the "variables" fixed property of Physics equations etc (which originally are are shortcuts for CUBA.VARIABLE with additional attribute being fixed on the metadata level, a note which did not transfer accurately enough probably to the specs here on github).

khiltunen:
I agree with @ahashibon . For example for Dirichlet condition we must specify the VARIABLE for which the condition is set. Otherwise it is not known.

stefanoborini:
> CUBA.VELOCITY is a CUBA.VARIABLE

there's nothing in the current format that allows this kind of specification.

stefanoborini:
@khiltunen the way this is normally done in the past is with the following writing

``variables: [CUBA.FOO, CUBA.BAR, CUBA.BAZ]``

khiltunen:
Yes that's true. I can see your point now.

docs/format.md
```diff
@@ -1,7 +1,7 @@
 Description of YAML format
 ==========================
 
-Specification version: 1.0
```
stefanoborini: There are no patchlevel versions to a format. You are changing something, it's either a minor increase (if compatible) or a major increase (if incompatible)

stefanoborini: Also, this is not the version of the _content_. It is the version of the format. The version of the content is the version of simphony-metadata.

yaml_files/simphony_metadata.yml
```diff
@@ -454,22 +454,24 @@ CUDS_KEYS:
...
     CUBA.NEUMANN:
 
   EMPTY:
-    definition: Empty boundary condition
-    parent: CUBA.VALUE_CONDITION
-    value: CUBA.NONE
+    definition: an entity to represent that no boundary condition is applied on that domain
+    parent: CUBA.CONDITION
+    CUBA.VALUE:
+      default: CUBA.NONE
```
stefanoborini: This is not consistent. CUBA.VALUE is not defined anywhere. It is not in cuba.yml. It is not in simphony_metadata.yml. 

If it were in cuba.yml, it would be a datatype and must therefore be int, string, double. None is not an option, first because we can't have a datatype of type None (it would not accept _anything_ else except None, which makes it quite useless)

if it were in simphony_metadata, then it would be a CUDS item, and by design the property must match the type, hence CUBA.NONE must be a subclass of CUBA.VALUE.

yaml_files/simphony_metadata.yml
```diff
@@ -454,22 +454,24 @@ CUDS_KEYS:
...
+      scope: CUBA.SYSTEM
```
stefanoborini: you can't have a CUBA.SYSTEM _and_ a default. It makes no sense. Either you have a default, or you let the compiler decide for it. They are mutually exclusive.

```
262     - ``default``:
263         Indicates the default value for the property once the ``CUDS Item`` has 
264         been instantiated.
265         The default MUST be type compatible with the property entry key
266         (eg. integers if the data is an integer)
267         If the key refers to a CUBA data, the default must match shape, type and length
268         requirements specified for the CUBA data, keeping into account the shape of the CUBA data
269         itself. 
270         If the key refers to a CUDS item, the default must belong to the hierarchy defined by
271         the CUDS item designated in the key.
272         if the ``scope`` is ``CUBA.SYSTEM``, this key MUST NOT be present

```

stefanoborini (changes requested):

yaml_files/cuba.yml
```diff
@@ -7,6 +7,11 @@ Purpose: CUBA provides the most primitive definitions used in the SimPhoNy metad
 
 CUBA_KEYS:
 
+  VALUE:
+    definition: a value that a variable or parameter takes, can be scalar or vector, double or integer, etc. The type is usually known from the context specific metadata
+    length:
+    type:
```
stefanoborini: type: any

yaml_files/cuba.yml
```diff
@@ -7,6 +7,11 @@ Purpose: CUBA provides the most primitive definitions used in the SimPhoNy metad
 
 CUBA_KEYS:
 
+  VALUE:
+    definition: a value that a variable or parameter takes, can be scalar or vector, double or integer, etc. The type is usually known from the context specific metadata
+    length:
```
stefanoborini: No length if type is any

yaml_files/simphony_metadata.yml
```diff
@@ -444,38 +443,34 @@ CUDS_KEYS:
...
-    definition: Dirichlet boundary condition, specify the value the solutions takes on the boundary of the domain
+  FIELD_CONDITION:
+    definition: Condition on (field) variables at domain boundaries encountered in the solution of partial differential equations (Direchlet, Neumann, Robin, etc)
     parent: CUBA.CONDITION
     models: [CUBA.CONTINUUM]
     CUBA.VARIABLE:
-      shape: (:)
-      default: []
+      shape: [1]
+    CUBA.VALUE:
```
stefanoborini: Considering that it's a very generic information at this level, wouldn't it make sense to have an H_VALUE on the ROBIN instead, and have H_VALUE defined as a shape 3 float array?

yaml_files/simphony_metadata.yml
```diff
@@ -503,32 +496,24 @@ CUDS_KEYS:
 
   INLET_OUTLET:
     definition: Inlet outlet boundary condition (outlet condition is zero gradient and inlet given variable value)
-    parent: CUBA.CONDITION
+    parent: CUBA.VALUE_CONDITION
```
stefanoborini: Do we have VALUE_CONDITION defined?

stefanoborini (changes requested):

stefanoborini:
In my opinion, and after analyzing the whole PR and discussing with Adham, my feeling is that:

1. Either you create specific CUDS for the conditions, and therefore have specific variable properties for them, something that is in the spirit of an ontology. (e.g. a constant pressure condition has a PRESSURE variable, a constant volume has a VOLUME variable)
2. or you define a generic condition (which is what is attempted here) and have yes, VARIABLE and VALUE. If that's the case, the following changes are required:
    1. VARIABLE must be a CUBA entry with type: cuba. Specs need to be expanded to cover this, and so is the compiler. Specifically, it allows only CUBA keys that are defined in cuba.yml.
    2. VALUE must be a CUBA entry with type: any. Specs need to be expanded to cover this, and so is the compiler. I would actually call it VARIABLE_VALUE
There is no need to define a type to describe what is in VALUE. It can be inferred from VARIABLE. 
 
   Additional requirements for the semantic correctness of the file is that if VARIABLE has a `shape` defined, VALUE must have the same shape and a default of compatible type (e.g. if VARIABLE is CUBA.PRESSURE, the VALUE default must not be a list of strings.)

With the second solution, you lose any information about the nature of the constraint. To know anything about it, you are forced to explore the specifics of what it contains. Unless you plan for hundreds of possible combinations leading to a proliferation of types, I think solution 1 is the most sensible, from a descriptiveness point of view.

yaml_files/simphony_metadata.yml
```diff
@@ -338,6 +338,13 @@ CUDS_KEYS:
     definition: Stress model from meso scopic to use in mixture model
     models: [CUBA.CONTINUUM]
 
+  PHASE_VOLUME_FRACTION:
+    definition: volume fractions of a phase (material) on a dataset entity
+    parent: CUBA.CUDS_ITEM
+    CUBA.MATERIAL:
+      shape: [1]
```
stefanoborini: this is the default. no need

yaml_files/simphony_metadata.yml
```diff
@@ -338,6 +338,13 @@ CUDS_KEYS:
...
+    CUBA.VOLUME_FRACTION:
+      shape: [1]
```
stefanoborini: same

yaml_files/simphony_metadata.yml
```diff
@@ -445,38 +452,71 @@ CUDS_KEYS:
...
+    definition: Dirichlet boundary condition to specify the value the solutions takes on the boundary of the domain.
     parent: CUBA.CONDITION
-    models: [CUBA.CONTINUUM]
-    CUBA.VARIABLE:
-      shape: (:)
-      default: []
     CUBA.MATERIAL:
-      shape: (:)
-      default: []
+      shape: [1]
```
stefanoborini: same

yaml_files/simphony_metadata.yml
```diff
@@ -445,38 +452,71 @@ CUDS_KEYS:
...
+
+  CONSTANT_VELOCITY_CONDITION:
+    definition: Constant velocity condition
+    parent: CUBA.DIRICHLET
+    models: [CUBA.CONTINUUM]
+    CUBA.VELOCITY:
+      shape: [1]
```
stefanoborini: same

yaml_files/simphony_metadata.yml
```diff
@@ -445,38 +452,71 @@ CUDS_KEYS:
...
+
+  CONSTANT_PRESSURE_CONDITION:
+    definition: Constant pressure condition
+    parent: CUBA.DIRICHLET
+    models: [CUBA.CONTINUUM]
+    CUBA.PRESSURE:
+      shape: [1]
```
stefanoborini: same

yaml_files/simphony_metadata.yml
```diff
@@ -445,38 +452,71 @@ CUDS_KEYS:
...
+
+  CONSTANT_VOLUME_FRACTION_CONDITION:
+    definition: Constant volume fraction condition
+    parent: CUBA.DIRICHLET
+    models: [CUBA.CONTINUUM]
+    CUBA.VOLUME_FRACTION:
+      shape: [1]
```
stefanoborini: etc.

yaml_files/simphony_metadata.yml
```diff
@@ -445,38 +452,71 @@ CUDS_KEYS:
...
+    definition: Zero gradient volume fraction condition
+    parent: CUBA.NEUMANN
+    models: [CUBA.CONTINUUM]
+    CUBA.VOLUME_FRACTION:
+      shape: [1]
+
+  ROBIN:
+    parent: CUBA.CONDITION
+    definition: A mixed boundary condition $\alpha \Phi (x) + \beta (x) \partial {\Phi} / \partial {\bf{n}} (x) = h(x)$, with $h$ is the value.
+    CUBA.DIRICHLET:
```
stefanoborini, line 507: if the default is a list, it probably needs a shape as well.

yaml_files/simphony_metadata.yml
```diff
@@ -445,38 +452,71 @@ CUDS_KEYS:
...
       default: []
-    CUBA.MATERIAL:
-      shape: (:)
+    CUBA.NEUMANN:
```
stefanoborini: Same.

stefanoborini (changes requested):

ahashibon:
@khiltunen @stefanoborini @mehdisadeghi This is still WIP but please check the current commits: 
1. each boundary condition is defined explicitly for each variable in the ontology/metadata
2. added cuds for the volume fractions. This needs one more step: 
   - currently a PHASE_VOLUME_FRACTION should be defined for each phase. 
   - each entitiy (cell, face etc) can then have: entitiy.data[PHASE_VOLUME_FRACTION] = [PHASE_VOLUME_FRACTION_for phase 1, PHASE_VOLUME_FRACTION for phase 2, etc]. 

the question is whether there is a way to force the value for entitiy.data[PHASE_VOLUME_FRACTION]  to be a list? or can we live with it as a convention for now?

stefanoborini:
@ahashibon 

```
CONSTANT_VOLUME_FRACTION_CONDITION:
    definition: Constant volume fraction condition
    parent: CUBA.DIRICHLET
    models: [CUBA.CONTINUUM]
    CUBA.VOLUME_FRACTION:
       shape: (:)
```

khiltunen:
Maybe it is enough to let the wrappers to check that the entity.data[PHASE_VOLUME_FRACTION] value is a list. Though it would be nice if there were a way to generally check this. It is the same for all entity data for example entity.data[CUBA.VELOCITY] should be a list of size 3.

ahashibon:
@stefanoborini can you please tell explain the issue with CONSTANT_VOLUME_FRACTION_CONDITION? 

@khiltunen this is true for all cuba.yml entities, so it should be a convention by now to do these checks!

stefanoborini:
@ahashibon that's how you express a list

yaml_files/simphony_metadata.yml
```diff
@@ -445,38 +452,71 @@ CUDS_KEYS:
...
+    parent: CUBA.NEUMANN
+    models: [CUBA.CONTINUUM]
+    CUBA.PRESSURE:
+      shape: [1]
+
+  ZERO_GRADIENT_VOLUME_FRACTION_CONDITION:
+    definition: Zero gradient volume fraction condition
+    parent: CUBA.NEUMANN
+    models: [CUBA.CONTINUUM]
+    CUBA.VOLUME_FRACTION:
```
stefanoborini, line 502: Shouldn't this be PHASE_VOLUME_FRACTION?

stefanoborini (changes requested):

ahashibon:
@stefanoborini This is true, and would have been nice if VOLUME_FRACTION was already containing the phase information, or if we have a strict ordering of the phases. The problem is that such a list does not convey information to which phase each item in the CUBA.VOLUME_FRACTION list refers to. This is why we created the CUDS item PHASE_VOLUME_FRACTION, to hold both the phase and the volume fraction of that phase. 

The alternative is to define in addition: 
PHASE_VOLUME_FRACTIONS:
    definition: volume fractions for a number of phases (material) on a dataset entity
    parent: CUBA.PHASE_VOLUME_FRACTION
    CUBA.PHASE_VOLUME_FRACTION:
      shape: (:)

(with plural S! or something similar), but this will complicate a bit the usage: 

entity.data[PHASE_VOLUME_FRACTIONS].PHASE_VOLUME_FRACTION[0].material and  entity.data[PHASE_VOLUME_FRACTIONS].PHASE_VOLUME_FRACTION[0].volume_fraction 

instead of
entity.data[PHASE_VOLUME_FRACTION].[0].material and  
entity.data[PHASE_VOLUME_FRACTION].[0].volume_fraction 

in both cases, the user/wrapper/visualisation tool etc, will definitely be able to infer/know to which phase (material) the index 0, etc, points too. 

in CFD we have usually 2 phases for which one volume fraction is sufficient (sum must be 1), but in general for N phases we need N-1 volume fractions. Having the material explicitly defined allows us much needed flexibility, for example we can define a volume fraction of the mixture material in this way: 

mixture=Material()
v=PHASE_VOLUME_FRACTION(material=mixture, volume_fraction=1)

it is up to the wrapper/user then of course to make sure all is consistent, CUDS only provides the efficient data structure, the wrapper the "reasonning".

@khiltunen  it is very tempting to also define: 

MIXTURE_MATERIAL:
  definition: a mixture of few phases or materials 
  parent: CUBA.MATERIAL 
  CUBA.MATERIAL: 
      shape: (:) 

note: if the above would preserve the order of the phases, we could then easilly live with just CUBA.VOLUME_FRACTION as a list, but the way proposed in this commit is more "safe"!

... comments are very much welcome!

stefanoborini:
What I am saying is that you defined PHASE_VOLUME_FRACTION and you never used it anywhere.

ahashibon:
(update) maybe I missuderstood : PHASE_VOLUME_FRACTION should be used directly on the entities, and it does not need to be used, for now, in the metadata it self. 

@stefanoborini we used it in the applications but not yet on any uploaded example. Also the lack of a clear way to relate it to a phase, meant we had to used ugly hacks! this is why we really need this urgently to make support for multiphase simulations possible in a standard clean manner!

@khiltunen please feel free to add your point of view ....

ahashibon:
@mehdisadeghi @stefanoborini can this be merged already?

stefanoborini:
@ahashibon There are review comments open and it needs merging of master.

yaml_files/simphony_metadata.yml
ahashibon: yes, good catch!

ahashibon: (I correct my previous reply!) good point but the material is already a property of the parent CUBA.NEUMANN and therefore the VOLUME_FRACTION is sufficient.

yaml_files/simphony_metadata.yml
ahashibon: Just one Dirichlet and one Neumann conditions per Robin.

yaml_files/simphony_metadata.yml
ahashibon: all done!

ahashibon:
@stefanoborini @kemattil please check, all review requests should be done!

yaml_files/simphony_metadata.yml
stefanoborini: Then there should not be a default

stefanoborini (changes requested):

stefanoborini (approved):

PR #123, opened 2017-03-14
Added none as an option for the entries in CUBA.

stefanoborini:
We need to handle None as well, and CUBA.NONE has a special type


ahashibon (commented):
for the time being (until we merge the basic simple types with the cuds structures), it looks ok. can be merged.

ahashibon (approved):
see previous comment.

